PR #23, opened 2011-07-19
Feature/loop syntax

wesen:
Out of the same clarity/optimization needs, I decided to use {{, }} to iterate (and only iterate) over an array (or a hash depending on the language), rather than the "i can do everything" {{#section}}.

let me know what you think


specs/loop.yml
```diff
@@ -0,0 +1,128 @@
...
+  - name: Nested loops single
+    desc: Loop over a hash should not push a new context
+    data:
+      a: 'first'
+      array:
+        a: 1
+        b: 2
+        c: 3
+    template: '{{,array}}{{a}} {{/array}}'
+    expected: 'first first first '
```
urzds: This test looks exactly like the previous one. Where is the difference?

urzds:
I also support this request, as I would like to iterate over a hash (key/value pairs). Could you please add an example of how that would look like? I propose `{{,hash}}key: {{key}}, value: {{value.nstd.idx}}{{/hash}}`

Using comma `,` as the "keyword" seems a bit unintuitive, but other than `*` I also have no good idea what else to use.

urzds:
What was the reason for closing this?

wesen:
Just closing after 5 years to clean up my open pull request view :)

urzds:
Shall I submit the patch in a new pull request?

PR #31, opened 2012-01-05
Add a test case checking context precedence

cjerdonek:
It doesn't look like the spec includes a test case checking that items higher in the context stack should take precedence -- especially a test case checking this in isolation.  Here is one.


Danappelxx:
Will be merged as #114, thank you for your contribution and sorry it took 9! years.

PR #35, opened 2012-01-08
Add a test that interpolated content should not be re-interpolated

cjerdonek:
This is basically the same as the test case of issue #32, but for basic (non-section) tags.


jgonggrijp (approved):
I'm not sure this test is strictly needed, but I do agree with it.

jgonggrijp:
Merging because it does not seem controversial.

PR #43, opened 2012-04-08
Add a test not to interpret strings as lists

cjerdonek:
In Python, for example (as in many other languages undoubtedly), strings are sequence types that share many of the same properties as lists and other iterator types:

  http://docs.python.org/library/stdtypes.html#sequence-types-str-unicode-list-tuple-bytearray-buffer-xrange

It would be good to add a test for this.  One does not seem to exist yet in the spec (incidental or not).

### DIFF
--- a/specs/sections.yml
+++ b/specs/sections.yml
@@ -98,6 +98,12 @@ tests:
     template: '"{{#list}}Yay lists!{{/list}}"'
     expected: '""'
 
+  - name: Strings Not Lists
+    desc: Strings should not be interpreted as list types.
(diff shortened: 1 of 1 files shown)


jgonggrijp:
Nowadays, the following test is included in the spec, which subtly implies that a string should not be iterated as a list:

https://github.com/mustache/spec/blob/5d3b58ea35ae309c40d7a8111bfedc4c5bcd43a6/specs/sections.yml#L68-L75

Hence, the proposed addition is not needed. Closing, discussion still welcome.

PR #76, opened 2014-06-16
Update interpolation.json to match .yml

softmoth:
Also update the Rakefile to ensure the `"__ATTN__"` notice is the
at the start of the file.


softmoth:
This pull request is obsoleted by #90.

PR #85, opened 2015-01-04
Add lambdas for Go

jabley:


jabley:
How can I help this progress?

jgonggrijp:
@jabley I'm going to help you get this merged. Would you like to update your branch first?

specs/~lambdas.yml
```diff
@@ -75,6 +78,7 @@ tests:
         clojure: '(def g (atom 0)) (fn [] (swap! g inc))'
         lisp:    '(let ((g 0)) (lambda () (incf g)))'
         pwsh:    'if (($null -eq $script:calls) -or ($script:calls -ge 3)){$script:calls=0}; ++$script:calls; $script:calls'
+        go:      'func() func()(string) { g := 0; return func() string { g++; return strconv.Itoa(g); } } ()'
```
jgonggrijp, line 81: Why convert the integer to string? If possible, I would prefer that you declare that the function returns an integer.

jabley: Requiring a string was a limitation of the mustache implementation that I originally wrote, but there's no reason to enforce that limitation, and I changed my version to happily handle other types.

jgonggrijp (approved):
Thanks!

PR #86, opened 2015-01-26
Adding functions for bash

fidian:
I did not rebuild the *.json files because the __tag__ => "code" bit is
missing whenever I run `rake build`.

Bash doesn't really support lambdas, but using "echo" and "$1" should be extremely easy for other tools to use.


jgonggrijp:
@fidian I'm going to help you get this merged. Would you like to update your branch first?

PR #89, opened 2015-08-08
Test for rendering nested partials.

Ryman:
I'm assuming this is expected behavior due to already having a test for recursive partials, and that the ruby implementation handles this behavior (from my limited testing).


jgonggrijp (approved):
I realize this is a rather late review, but I'm in favor of it. 👍 

@spullara or @bobthecow, any comments?

spullara:
Makes sense to add a test for this IMHO.

PR #90, opened 2015-08-28
Ensure json files are formatted properly

softmoth:
The JSON spec files should end with a newline. Also, the "**ATTN**"
notice should be the first entry, not randomly placed, so it is
visible even in large spec files.


softmoth:
This pull request is obsoleted by #113.

Issue #52, opened 2012-07-20
Feature: moar dot notation hotness


bobthecow:
1. Given that `{{ . }}` resolves as the current top of the context stack;
2. And when any falsey segment in a dotted name is encountered, the whole name yields `''`;
3. A name like `{{ .name }}` should imply `{{ [current scope].name }}`;
4. Thus, it must resolve as truthy _only_ if a member of the current scope matches `name`.

This is a one-liner to implement in Mustache.php, and seems completely in line with current spec, so I'm considering adding it. Can we get this into spec v1.2?

c.f. https://github.com/bobthecow/mustache.php/issues/98

tomekwi:
Bump! That’s a very useful feature – see <https://github.com/mustache/spec/pull/92#issue-104010862>.

aubelsb2:
This is what I tried (failed) and need.

jgonggrijp:
I reopened #92. Please continue discussion about the proposal over there, or discuss stack masking in general in #154.

PR #92, opened 2015-08-31
Add spec for leading dot

tomekwi:
Fixes https://github.com/mustache/spec/issues/52

The feature proposed there is extremely useful. It’s the only way to DRY out templates which deal with recursive data (like here: <http://stackoverflow.com/a/13592747/2816199>). The new syntax should also allow more radical perf optimizations.


tomekwi:
By the way, this PR is just an application of [the original diff](https://github.com/mustache/spec/issues/52#issuecomment-7123144) by @bobthecow.

tomekwi:
Closing to remove clutter 🙂

Feel free to open if you think this is mergeable.

PR #93, opened 2015-09-18
Enhance the Standalone Indentation test

Ryman:
The indentation should only be applied at the start of a line.

### DIFF
--- a/specs/partials.yml
+++ b/specs/partials.yml
@@ -80,7 +80,7 @@ tests:
     expected: ">\n  >\n  >"
 
   - name: Standalone Indentation
-    desc: Each line of the partial should be indented before rendering.
+    desc: Each new line of the partial should be indented before rendering.
     data: { content: "<\n->" }
     template: |
       \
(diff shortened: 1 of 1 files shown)


jgonggrijp:
Does this change actually solve anything? I mean, I don't think you can pass the new version of the test without also passing the old version.

If it does solve something, I would prefer the new version to coexist alongside the old version.

specs/partials.yml
```diff
@@ -106,6 +106,27 @@ tests:
        |
       /
 
+  - name: Standalone Indentation
```
jgonggrijp, line 109: Please add something to this `name` to distinguish it from the previous test.

specs/partials.yml
```diff
@@ -106,6 +106,27 @@ tests:
...
+    desc: Indentation should only be applied after each newline in a partial.
```
jgonggrijp, line 110: This description is not strictly true, because the indentation should also be applied to the first line, before any newline.

jgonggrijp (changes requested):
Thank you for separating this from the existing test. With some slight changes (see comments), I think we can accept this. Would you be able to update the JSON as well?

@spullara or @bobthecow, please take a quick look if you can.

Ryman:
@jgonggrijp updated and re-built the json files

jgonggrijp (approved):

PR #95, opened 2015-10-06
Add list context check

andrewthad:
Added a test case to make sure that hashmaps don't get dropped when you iterate over a list.


Danappelxx:
Will be merged as part of #114, thank you for your contribution!

PR #113, opened 2020-12-24
Refresh the JSON specs

gasche:
This PR contains two changes:
- A minor change to the Rakebuild file that was necessary, on my machine, to get `rake build` to render JSON specs properly (otherwise the ATTN warning would be at the end of each test, instead of at the begining as in their current versioned form.)
- A new run of `rake build` that refreshes .json files that are stale compared to the .yml reference.


gasche:
I now realize that this PR overlaps with #90 from August 2015 which is doing the same thing. This should have been fixed long ago.

spullara:
Any reason not to merge this?

gasche:
I would be in favor of merging, indeed!

I don't think it is necessary, but before merging you could maybe try to see if you can reproduce the warning-placement issue on your system: if you run the command to refresh the specs (`rake build`) on the current main branch, do you see the warning as the first key of the generated JSON objects, as intended, or do you observe a difference placement as on my machine?

Danappelxx:
Confirming the warning-placement issue, thanks for fixing. Agree, should have been merged a long time ago.

PR #114, opened 2020-12-24
Clarify context stacks

gasche:
Some aspects of context stacks are not properly exercised by the current spec -- I realized this when hitting a bug in an implementation that was checked against the spec. When entering a new context, previous contexts should remain available for variables names that are not defined in the new contexts; this was only tested for object sections, not for list sections or non-false scalar sections.

In particular, inside a non-false scalar section `{{#name}}...{{/name}}` (whose value is neither a list nor a string), `{{.}}` and `{{name}}` should always be equivalent.


gasche:
After preparing this PR I realized that fairly old PRs lying around unanswered are trying to cover the same hole in the specification: #31 from January 2012, and #95 from October 2015. The new tests that I am proposing supersede #31, but not #95, so I just included the commit of #95 in the current PR.

Who is maintaining this repository these days? (cc @locks, who is the last to have a merged a Pull Request.) These changes should be uncontroversial as they are not adding new features, just improving the testing coverage for existing features, and it is unfortunate that they were delayed for so long.

gasche:
cc @spullara , @Danappelxx : I think this may be also an innocuous change worth merging.

Danappelxx:
Thank you for the detailed commit message. Can you re-run `rake build`? Will go ahead and merge afterwards.

gasche:
Thanks! I rebased on top of master and ran `rake build` to keep the JSON specs up-to-date.

PR #91, opened 2015-08-28
Add perl6 lambda support

softmoth:
This request obsoletes pull request #78, and is built on top of pull request #90.

PR #116, opened 2021-03-18
~lambdas: Add Raku language tests (formerly Perl 6)

softmoth:
Replaces #91. In the interim since that pull request was created, Perl 6 was renamed to Raku.


Danappelxx:
I don't know perl 6 very well, so unfortunately I can't review this PR for correctness. If someone else can review this, would be great.

Alternatively, we can wait until we have CI set up which can actually test the spec against your implementation.

softmoth:
Thanks, Danappelxx. Aside from the `perl6` to `raku` rename, this is the same exact lambda tests that have been [running on raku-Template-Mustache][raku-mustache-ci] for years. So it is safe to use. I will push an update to raku-Template-Mustache to refer to the official repo w/ the `raku` naming as soon as I fix the `sections.yml` tests, which have two new checks that raku-Template-Mustache doesn't pass. But that should be considered independent of this PR.

[raku-mustache-ci]: https://travis-ci.org/github/softmoth/raku-Template-Mustache/jobs/711659653#L546

Thank you!

Tim (aka softmoth)

softmoth:
I've pushed a new version that fixes the bug exposed by #114 (great to have new test cases) and includes this PR's version of `~lambdas.json`. You can verify its success at https://travis-ci.org/softmoth/raku-Template-Mustache . Again, this has been tested thousands of times over the years by the Perl 6 user base; the 'perl6' to 'raku' change is the only new thing. I hope that helps.

Kind regards,

Tim

Danappelxx:
Fantastic! Thank you for your contribution.

PR #119, opened 2021-03-19
Pretty print JSON files, and terminate with newline

softmoth:
Text files that don't end with a newline can be troublesome or unacceptable in some contexts.

The only down side is a bit of merging hassle for existing pull requests. But that happens anyways, and of course the YAML file is authoritative so JSON conflicts can be ignored in practice.


gasche:
Looks fine to me.

(It would also be nice to have some kind of human-friendlier printing of the generated JSON, because currently it is painful for implementations that use the JSON to locate the source of a test failure within the big generated one-liner.)

softmoth:
Yes, @gasche, good point. I found how to do that. Its default looks pretty good, I think.

gasche:
Very nice! Thanks.

Danappelxx:
I'm a tiny bit hesitant to merge this since all other pull requests would need to be re-run. I'll merge a few and then come back to this one.

Danappelxx:
I am very much in favor of the quality-of-life and readability improvement this provides! I merged a few active pull requests, can you go ahead and re-generate the json?

softmoth:
I force-pushed an update of the `specs/*.json` files.

Danappelxx:
Thanks!

PR #120, opened 2021-03-19
Remove inaccurate documentation about __tag__ in JSON specs

softmoth:
There is no special `__tag__` key in the JSON files. Instead, it
is the `lambda` entry which is significant. While Ruby may depend
on the YAML `!code` tag, that doesn't make it into the JSON format.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,12 +32,6 @@ best-effort basis.
 These should be identical to the YAML specifications, but if you find the need
 to regenerate them, they can be trivially rebuilt by invoking `rake build`.
 
-It is also worth noting that some specifications (notably, the lambda module)
-rely on YAML "tags" to denote special types of data (e.g. source code).  Since
-JSON offers no way to denote this, a special key ("`__tag__`") is injected
-with the name of the tag as its value.  See `TESTING.md` for more information
-about handling tagged data.
-
 Optional Modules
 ----------------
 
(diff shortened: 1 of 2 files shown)


jgonggrijp (changes requested):
I'm not opposed to using backticks instead of quotes, but I don't think that the removal of the nesting in the `TESTING.md` is an improvement. Those steps are actually substeps of the top-level step 2.

More importantly, historically it is not true that the `__tag__` isn't there. On latest `master` it seems to have disappeared, but it was still there in the 1.1.3 tag. I would rather consider this a regression and fix it than changing the README because of it.

softmoth:
I see. The `__tag__` got removed after #113 was merged (commit 9944f2). I will look into how to put it back.

The nesting isn't valid Markdown, and breaks the display when viewing the rendered file, as on Github. Please compare:
- Old: https://github.com/mustache/spec/blob/master/TESTING.md
  - Numbering goes 1, 2, 3 (wrong), 4 (wrong), 2.1 (out-dented, not part of the list), 3, 4.
  - Bulleted list under section 2.1 is entirely preformatted, rather than rendered as the intended list items
- New: https://github.com/softmoth/mustache-spec/blob/readme-yaml-notag/TESTING.md

I will moving what was "item" 2.1 back under item 2 as a mock sub-heading to see if it is better. Unfortunately, Markdown doesn't handle this particular structure elegantly as far as I know. I'll open a separate issue just for fixing this display issue.

Thanks for your review!

softmoth:
BTW, the Rakefile hasn't inserted `__tag__` since Ruby 1.9.3 switched the YAML processor from Syck to Psych. It's possibly why the JSON files hadn't been updated since 2011.

jgonggrijp:
@softmoth You are right that the current numbering layout is broken. I didn't realize that.

PR #121, opened 2021-03-20
Fix Rakefile to add __tag__: code for ~lambda test data 

softmoth:
_Note: this PR is built on top of #119. Please just look at the 2nd commit, the first is from #119._

This was broken when Ruby switched YAML engines from Syck to Psych
around Ruby 1.9.3. Syck was removed completely around version 2.0.
The code didn't fail, because Psych also implements the add_builtin_type
routine, but it does not recognize `!code` as a builtin type (because it
isn't; builtins look like `!!str` or `!!map`).

Instead, Psych requires a custom class to handle decoding the tagged
YAML item to Ruby.

And then, to get JSON to actually display the object, it is easiest to
just inherit from `Hash`.

_I am not a Ruby developer. I welcome any code changes that would improve this._


softmoth:
I should probably add just a few notes about this for posterity, since there's zero documentation for Psych's `add_tag`.

Learning resources:
- a blog post by @djellemah was particularly helpful: http://djellemah.com/blog/2014/07/23/yaml-deserialisation/
- a stackoverflow answer also helped: https://stackoverflow.com/questions/11239884/ruby-yaml-custom-domain-type-does-not-keep-class

I tried to `def initialize(tag_name)` and then call `add_tag('code', TaggedMap.new('code')`, and `def yaml_tag: '!' + @tag`, but couldn't find a combo that works, and it would have made the code a lot harder to read without providing any real benefits. So I left the `!code` tag name hard-coded.

I'd rather not have inherited from `Hash`, but I could not get `as_json` to be called properly on the object. Calling `doc.as_json()` fails since `Hash` doesn't implement it, and it makes no sense to try to do so for this use case. There may well be a better solution for this, but I couldn't find anything. Anyways, I doubt that any proper solution would be as concise and legible, so perhaps it's best as is.

jgonggrijp (approved):
I'm not a Ruby coder, either, but the change looks very reasonable to me, at least in terms of the amount of code added. Also highly valuable to have these tags back in the JSON.

Thanks a lot @softmoth!

softmoth:
I've force-pushed an update of the `specs/*.json` to match the latest revision of #119.

Danappelxx:
Thanks!

PR #122, opened 2021-03-20
Fix TESTING.md Markdown list syntax; no content changes

softmoth:
The present indentation isn't valid Markdown, and breaks the display
when viewing the rendered file, as on Github. Please compare:

- Old: https://github.com/mustache/spec/blob/f75ffbf0a90e6ccecb11df58cdb4cc400ac2919d/TESTING.md
  - Numbering goes 1, 2, 3 (wrong), 4 (wrong), 2.1 (out-dented, not part
    of the list), 3, 4.
  - Bulleted list under section 2.1 is entirely preformatted, rather
    than rendered as the intended list items

- New: https://github.com/softmoth/mustache-spec/blob/format-testing-doc/TESTING.md

This patch moves what was "item" 2.1 back under item 2 without an extra
list item. Unfortunately, Markdown doesn't handle this particular
structure elegantly as far as I know.

It also switches to backtick quotes for all literals.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -6,41 +6,40 @@ In general, the process for each `.yml` file is as follows:
 
 1. Use a YAML parser to load the file.
 
-2. For each test in the 'tests' array:
+2. For each test in the `tests` array:
 
-  1. Ensure that each element of the 'partials' hash (if it exists) is
-     stored in a place where the interpreter will look for it.
+    1. Ensure that each element of the `partials` hash (if it exists) is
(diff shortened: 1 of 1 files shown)


jgonggrijp:
This looks good to me, but I suspect that the top-level items 3 and 4 (in your new version) were also supposed to be subitems of the top-level item 2. What do you think?

softmoth:
Yes, you're right. Commit amended.

jgonggrijp (approved):

Issue #101, opened 2016-11-20
`null` interpolation


berdario:
Which alternative should be part of the spec?

```
  - data: {nullkey: null }
    template: "{{nullkey}}"
    expected: ""
```

```
  - data: {nullkey: null }
    template: "{{nullkey}}"
    expected: "null"
```

I've seen a Mustache implementation that yields the latter, but then I checked [the demo](https://mustache.github.io/#demo), and its behavior instead was the former

groue:
What do you think?

berdario:
Me?

I think that the former behavior might be less surprising, but I don't have a strong opinion.

I don't think that this should be left unspecified

groue:
This repository has not been updated in years.

berdario:
Yes, about 1year and a half, but since the Ruby implementation repository is still being worked on, and that whoever has write rights on that repo should also have them on this one, there's no excuse for this situation to persist (I know that you're not part of the organization, so I'm not blaming you)

~~Either they pick up the burden, or they give access to others to this repo and the [website](https://github.com/mustache/mustache.github.com) repo or, as a third alternative, if they don't want to maintain the spec anymore they should declare it an unmaintained/deprecated, give up any pretense of being a complete spec and let other standards flourish~~

Ok, I strikethrough'ed the text since I'm afraid it could've come off as a bit bitter... also I found this old blog post: http://writing.jan.io/mustache-2.0.html So the idea of a Mustache2 is not new, and Handlebars apparently never went anywhere (outside of javascript and few languages)

I guess that I should get used to the idea that good and complete standards are an exception, not the rule

groue:
Yes, this is very frustrating. You may be interested in https://github.com/groue/GRMustacheSpec, an extended Mustache test suite (which addresses the [rendering of null](https://github.com/groue/GRMustacheSpec/blob/master/Tests/general/variables.json#L65))

Danappelxx:
Reviving! We have a couple maintainers now... Personally I think rendering `"null"` is totally incorrect. Would appreciate a pull request adding this to the spec.

PR #123, opened 2021-04-18
Specify how to interpolate null (#101)

jgonggrijp:
See discussion in #101.

### DIFF
--- a/specs/interpolation.json
+++ b/specs/interpolation.json
@@ -101,6 +101,33 @@
       "expected": "\"1.21 jiggawatts!\""
     },
     {
+      "name": "Basic Null Interpolation",
+      "desc": "Nulls should interpolate as the empty string.",
+      "data": {
+        "cannot": null
+      },
+      "template": "I ({{cannot}}) be seen!",
+      "expected": "I () be seen!"
+    },
+    {
+      "name": "Triple Mustache Null Interpolation",
+      "desc": "Nulls should interpolate as the empty string.",
+      "data": {
+        "cannot": null
+      },
+      "template": "I ({{{cannot}}}) be seen!",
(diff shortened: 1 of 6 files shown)


gasche:
Indeed, interpolating NULL to the empty string reflects the consensus in #101 (among a small group of participants), and is highly consistent (unlike "null") with the following existing parts of the interpolation specification:

> If any part fails resolution, the result should be considered falsey, and should interpolate as the empty string.
> Data should be coerced into a string (and escaped, if appropriate) before interpolation.

I'm in favor of merging this change.

Danappelxx:
Agreed, seems like an unfortunate oversight. Does anyone know of a Mustache implementation that interpolates `null` to something other than an empty string? I know it was mentioned in #101 but the poster did not specify which implementation.

As a side note, I think it would be beneficial to add a test for [null being falsey](https://github.com/groue/GRMustacheSpec/blob/master/Tests/values/null.json#L15-L20).

jgonggrijp:
> As a side note, I think it would be beneficial to add a test for [null being falsey](https://github.com/groue/GRMustacheSpec/blob/master/Tests/values/null.json#L15-L20).

Oh, right. Shall I add that in here?

Danappelxx:
I'd consider it a part of "interpolating null", so yeah throw the test case in this pr.

jgonggrijp:
There you go!

Would you mind to review https://github.com/mustache/mustache/pull/266?

jgonggrijp:
bump!

Danappelxx:
Reviewed! Sorry for the delay.

Danappelxx:
Merged & tagged in `v1.2.1`

jgonggrijp:
Thanks!

Issue #82, opened 2014-12-21
Missing test: triple mustache support for implicite iterator


kanru:
Found by a user of cl-mustache: https://github.com/kanru/cl-mustache/issues/18

The spec test doesn't have tests for `{{.}}` and `{{{.}}}` with HTML escaping characters.

Danappelxx:
Feel free to submit a pull request - will be merged by new maintainers.

jgonggrijp:
This one can be closed, as it was resolved in the PR referenced above.

kanru:
Thanks!

jgonggrijp:
Welcome!

PR #124, opened 2021-04-19
Specify interpolation behavior for implicit iterator

jgonggrijp:
My follow-up on #82. Although I don't think my suggested changes would be very controversial, there is some slight potential for breakage here. While my own implementation passes the new tests without modifications, I imagine that some implementations might support the implicit iterator only within a section. I'm open to discussion.

### DIFF
--- a/specs/interpolation.json
+++ b/specs/interpolation.json
@@ -245,6 +245,41 @@
       "expected": ""
     },
     {
+      "name": "Implicit Iterators - Basic Interpolation",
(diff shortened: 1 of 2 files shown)


Danappelxx:
Seems like a no-brainer, I would assume that most implementations already do this and that this was in the original intent of the spec.

I'll wait a few days just in came someone drops by to tell me I'm wrong, though.

jgonggrijp:
bump!

Danappelxx:
Merged & tagged in `v1.2.1`

PR #126, opened 2021-05-06
Added Lambdas for PowerShell/PSMustache

splumhoff:
Added Lambda-Definitions for PowerShell / PSMustache.

### DIFF
--- a/specs/~lambdas.json
+++ b/specs/~lambdas.json
@@ -15,7 +15,8 @@
           "php": "return \"world\";",
           "python": "lambda: \"world\"",
           "clojure": "(fn [] \"world\")",
-          "lisp": "(lambda () \"world\")"
+          "lisp": "(lambda () \"world\")",
+          "pwsh": "\"world\""
         }
       },
       "template": "Hello, {{lambda}}!",
@@ -35,7 +36,8 @@
           "php": "return \"{{planet}}\";",
           "python": "lambda: \"{{planet}}\"",
           "clojure": "(fn [] \"{{planet}}\")",
-          "lisp": "(lambda () \"{{planet}}\")"
+          "lisp": "(lambda () \"{{planet}}\")",
+          "pwsh": "\"{{planet}}\""
         }
       },
       "template": "Hello, {{lambda}}!",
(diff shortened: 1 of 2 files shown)


specs/~lambdas.json
```diff
@@ -15,7 +15,8 @@
           "php": "return \"world\";",
           "python": "lambda: \"world\"",
           "clojure": "(fn [] \"world\")",
-          "lisp": "(lambda () \"world\")"
+          "lisp": "(lambda () \"world\")",
+          "pwsh": "\"world\""
```
jgonggrijp, line 19: Is this really PowerShell's equivalent of a lambda? Shouldn't there be something like a `function` around the expressions? Or perhaps `echo` or `return` in front?

splumhoff: Hi Julian,

no, it's not necessary. Unlike other languages PowerShell returns the output of each expression, when the output is not captures in any way e.g. with a variable assignment. The return statement is optional in this case.

You can find more about this behavior here https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_return?view=powershell-7.1.

PowerShell does not have lambda functions in the way you may know them, but the nearest equivalent are ScriptBlocks which works similar. (https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_script_blocks?view=powershell-7.1)

~~~PowerShell
PS> $foo = { "Hello" }
PS> Invoke-Command -ScriptBlock $foo
Hello

PS> $foo = { $args[0] + 1 }
PS> Invoke-Command -ScriptBlock $foo -ArgumentList 1
2
~~~

jgonggrijp: So in this specific case, `lambda` will behave as a property that dynamically returns `"world"` (when invoked as a function/method) rather than just acting as a property that has this value statically?

splumhoff (commented):
@jgonggrijp Do you need any further information regarding the PR? :)
I'd release the new Version of PSMustache 1.3 with lambda support and it would be great if the tests were in the official repo at that time.

splumhoff:
Dear all,

any chance to get this PR merged? 
@Danappelxx I saw that you merged the previous PR. If you have any questions, please don't hesitate to ask. :)

Danappelxx:
Hi! Sorry, I just really don't know Powershell so I can't give this a thorough review. I guess regardless there's no harm done (until the next Powershell implementation comes around, anyway) so I'll go ahead and merge now. Thank you for your contribution!

PR #131, opened 2021-07-13
Additional inheritance specs for block reindentation

### LINKED ISSUE #130: Proposal on how to specify indentation of blocks
In #75 and #125, the inheritance spec was discussed at length. An issue that has been discussed, but undecided, is how it interacts with the notions of standalone tags and indentation.

## Setting the stage

Currently, it is specified that when a standalone partial tag is indented, the entire corresponding partial template will be indented by the same amount before rendering and interpolation.

_template.mustache_

```
    {{>partial}}
```

_partial.mustache_

```
Hello there!
```

_output_

```
    Hello there!
```

*Try the above example in the [playground](https://jgonggrijp.gitlab.io/wontache/playground.html) by pasting the following code:*

```
{"data":{"text":"null"},"templates":[{"name":"template","text":"    {{>partial}}"},{"name":"partial","text":"Hello there!"}]}
```

A parent tag is essentially a partial tag with a matching closing tag, allowing it to have internal structure.

_template.mustache_

```
{{<partial}}{{/partial}}
```

_output_

```
Hello there!
```

*Try the above example in the [playground](https://jgonggrijp.gitlab.io/wontache/playground.html) by pasting the following code:*

```
{"data":{"text":"null"},"templates":[{"name":"template","text":"{{<partial}}{{/partial}}"},{"name":"partial","text":"Hello there!"}]}
```

A block tag essentially enables us to define and expand an embedded partial template inside a larger template. Such a block can be passed to a parent template.

_template.mustache_

```
{{<introduction}}{{$title}}Professor{{/title}}{{/introduction}}
```

_introduction.mustache_

```
I'm {{$title}}Darth{{/title}} {{$name}}Vader{{/name}}.
```

_output_

```
I'm Professor Vader.
```

*Try the above example in the [playground](https://jgonggrijp.gitlab.io/wontache/playground.html) by pasting the following code:*

```
{"data":{"text":"null"},"templates":[{"name":"template","text":"{{<introduction}}{{$title}}Professor{{/title}}{{/introduction}}"},{"name":"introduction","text":"I'm {{$title}}Darth{{/title}} {{$name}}Vader{{/name}}."}]}
```

From the recursion spec, it follows that blocks are passed across multiple inheritance levels even when intermediate levels leave the block implicit. While not directly relevant to the current issue, I will be exploiting this in some of my examples, because it enables me to separate the indentation of the parent tag from the indentation of the block tags on both ends.

_template.mustache_

```
{{<intermediate}}{{$name}}Moder{{/name}}{{/intermediate}}
```

_intermediate.mustache_

```
{{<introduction}}{{/introduction}}
```

_output_

```
I'm Darth Moder.
```

*Try the above example in the [playground](https://jgonggrijp.gitlab.io/wontache/playground.html) by pasting the following code:*

```
{"data":{"text":"null"},"templates":[{"name":"template","text":"{{<intermediate}}{{$name}}Moder{{/name}}{{/intermediate}}"},{"name":"introduction","text":"I'm {{$title}}Darth{{/title}} {{$name}}Vader{{/name}}."},{"name":"intermediate","text":"{{<introduction}}{{/introduction}}"}]}
```

## The problem

In the previous section, I intentionally omitted indentation in the demonstration of parent and block tags, because there is currently no specification of how that should behave. Since a parent tag pair without blocks is basically just a partial, the following interpretation would probably be uncontroversial:

_template.mustache_

```
    {{<partial}}{{/partial}}
```

_output_

```
    Hello there!
```

Adding blocks to the mix, most people will probably also agree with the whitespace in the following example.

_template.mustache_

```
{{<intermediate}}
{{$greeting}}
high five
{{/greeting}}
{{/intermediate}}
```

_intermediate.mustache_

```
Hi,
    {{<invitation}}{{/invitation}}
```

_invitation.mustache_

```
please give me a:
    {{$greeting}}
    hug
    {{/greeting}}
```

_output_

```
Hi,
    please give me a:
        high five
```

The above examples are relatively clear-cut because all tags are standalone and the block is only indented in the place of expansion (in invitation.mustache). Things become fuzzier when indentation appears in more places,

_template.mustache_

```
{{<intermediate}}
    {{$greeting}}
        high five
    {{/greeting}}
{{/intermediate}}
```

_invitation.mustache_

```
please give me a:
    {{$greeting}}
        hug
    {{/greeting}}
```

or when some tags don't clear on both sides.

_template.mustache_

```
{{<intermediate}}{{$greeting}}
high five
{{/greeting}}{{/intermediate}}
```

_invitation.mustache_

```
please give me a:
    {{$greeting}}hug{{/greeting}}
```

Hence, in order to promote consistent behavior across different Mustache implementations, I'm suggesting to specify how template inheritance and whitespace should interact (credits to @gasche for first suggesting this). I'll open the discussion with my current ideas of what might be the best way to approach this. Before I do that, though, let me introduce some concepts that I think are relevant.

### Template space vs. comment space

Comment tags introduce "dead" space within a template that is never rendered.

```
before {{! never rendered }} after
```

To highlight this distinction, let's paint the comment space black.

```
before {{!████████████████}} after
```

Per the spec, the contents within a pair of parent open/close tags default to being comment space, too.

```
{{<parent}} never rendered {{interpolation}} {{/parent}}
```
```
{{<parent█████████████████████████████████████████████}}
```

This also visually highlights the similarity between a parent and a partial.

```
{{>parent}}
```

By way of exception, block tags can "punch holes" in the comment space within parent tag pairs, creating what could be considered "template islands", with "comment barriers" in between them.

```
a {{<parent}} b {{$block}} c {{/block}} d {{/parent}} e
```
```
a {{<parent███████$block}} c {{████████████████████}} e
```

Within a regular, continuous stretch of template space, everything is linearly ordered. As a result, everything that comes before a particular tag will directly influence the position of everything that comes after that tag.

```
If this wasn't here... {{#section}} ... this would be more to the left. {{/section}}
```

Comment barriers provide a visual way to remind us that there is no such direct relationship within and across parent tags. Blocks need not appear in the same order within a parent tag pair as within the corresponding parent template, or might even be ignored. Also, a block may be significant within a parent tag pair, but play only a minor role within the parent template.

```
{{<parent████$block}} If this wasn't here... {{███████████████████}} ... this might as well be just as much to the left.
```

We could say that the larger template outside of the parent tag pair and the template islands within them act like "parallel worlds" that don't really interact, until everything "lands" in the parent template.

### Parameter blocks vs. argument blocks

A block *outside* any parent tag pair marks a place in the template that can be configured. It acts as a *parameter* of the template. If not configured from the outside, its contents serve as the default.

```
{{$parameter}} default {{/parameter}}
```

A block *inside* a parent tag pair provides the external configuration for such a parameter. It acts as an *argument* for another template.

```
{{<parent████$parameter}} argument {{███████████████████████}}
```

### Definition vs. expansion

As I mentioned before, the contents of a block can be thought of as an embedded partial template. To illustrate, compare the following pair of templates:

_block.mustache_

```
{{<parent████$block}}Hello there!{{█████████████████}}
```

_parent.mustache_

```
{{$block}}{{/block}}
```

with the following:

_block.mustache_

```
Hello there!
```

_parent.mustache_

```
{{>block}}
```

The output is the same in both cases (`Hello there!`) and in both cases, content from one template is inserted in the other. In both cases, this content is *defined* inside block.mustache while parent.mustache is the place where it is *expanded*. The main differences are (1) which template is "outermost" and (2) how much of a template file is dedicated to the definition of the "transplanted" content. In partials, the definition is always the entire file, while in blocks, it never is.

It is important to distinguish definition from expansion, because the site of expansion can be identical to the site of definition, which might hide the distinction. In blocks, this happens when a parameter block is never configured from the outside and the default ends up being rendered.

```
{{$block}}Hello there!{{/block}}
```

There are several asymmetries between block definitions and expansions, and between parameter and argument blocks. A parameter block is a site of both definition and expansion, while an argument block is only a site of definition. Every parameter block ends up being expanded into, but not every defined block is expanded. A block defined in argument position can only be expanded elsewhere, while a block defined in parameter position can never be expanded elsewhere.

### Clearance

For most tag types, it is specified that they should not introduce blank lines just by existing. This is why

```
a
{{!████}} b
c
```

renders into

```
a
 b
c
```

while

```
a
{{!████}}
c
```

renders without the middle line.

```
a
c
```

We say that the latter comment is "standalone" while the former is not. For those tag types where it is specified unambiguously, standalone always means that there is *start clearance*, i.e., no content except for whitespace between the preceding line start and the opening delimiter (`{{`) of the tag, as well as *end clearance*, no content expect for whitespace between the closing delimiter (`}}`) of the tag and the following line end.

For parent and block tag pairs, a more nuanced concept of standalone is likely necessary, due to the previous distinctions between template and comment space, parameter and argument blocks, and definition and expansion. Let's first consider the state space for a single tag (`{{!}}` below). There are four possible combinations depending on whether it clears on the start and end or not:

```
┌───────────┬─────────────────────┐
│ start     │    end clearance    │
│ clearance │   yes   ╷    no     │
│      ╶────┼─────────┼───────────┤
│       yes │ a       │ a         │
│           │ {{!}}   │ {{!}} b   │
│           │ b       │           │
│      ╶────┼─────────┼───────────┤
│        no │ a {{!}} │ a {{!}} b │
│           │ b       │           │
└───────────┴─────────┴───────────┘
```

In the original tags such as section, comment and change delimiter, we consider the yes/yes case in the top left as standalone and the other three as "not standalone". In order to distinguish the other three cases from each other as well, let's name all combinations with a letter: **B** when the tag clears on **b**oth delimiters, **S** when it clears only at the **s**tart, **E** when it clears only at the **e**nd and **N** when it clears on **n**either delimiter.

In blocks, the opening and closing tags may independently have each of the four types of clearance, so we end up with sixteen possible combinations for the block as a whole:

```
┌─────────┬─────────────────────────────────────────────────────────────┐
│ opening │                     closing tag                             │
│ tag     │     B     ╷     S     ╷        E        ╷        N          │
│       ╶─┼───────────┼───────────┼─────────────────┼───────────────────┤
│       B │ a         │ a         │ a               │ a                 │
│         │ {{$}}     │ {{$}}     │ {{$}}           │ {{$}}             │
│         │ b         │ b         │ b {{/}}         │ b {{/}} c         │
│         │ {{/}}     │ {{/}} c   │ c               │                   │
│         │ c         │           │                 │                   │
│       ╶─┼───────────┼───────────┼─────────────────┼───────────────────┤
│       S │ a         │ a         │ a               │ a                 │
│         │ {{$}} b   │ {{$}} b   │ {{$}} b {{/}}   │ {{$}} b {{/}} c   │
│         │ {{/}}     │ {{/}} c   │ c               │                   │
│         │ c         │           │                 │                   │
│       ╶─┼───────────┼───────────┼─────────────────┼───────────────────┤
│       E │ a {{$}}   │ a {{$}}   │ a {{$}}         │ a {{$}}           │
│         │ b         │ b         │ b {{/}}         │ b {{/}} c         │
│         │ {{/}}     │ {{/}} c   │ c               │                   │
│         │ c         │           │                 │                   │
│       ╶─┼───────────┼───────────┼─────────────────┼───────────────────┤
│       N │ a {{$}} b │ a {{$}} b │ a {{$}} b {{/}} │ a {{$}} b {{/}} c │
│         │ {{/}}     │ {{/}} c   │ c               │                   │
│         │ c         │           │                 │                   │
└─────────┴───────────┴───────────┴─────────────────┴───────────────────┘
```

We can identify each combination by concatenating the letters for the opening and closing tags, so from left to right, the combinations at the top row would be named **BB**, **BS**, **BE** and **BN**. If we want to promote consistent behavior across implementations, the specification will need to address unambiguously what should happen with the whitespace in all sixteen combinations.

## Proposal

I suggest a small set of rules, which should unambiguously specify what happens with whitespace in all cases and which I hope will also be acceptable to most people. The following rules were in part inspired by ideas previously shared by @gasche.

### 0. Standalone tags

Individual block and parent tags should not introduce blank lines in the output when they clear on both delimiters. I'm numbering this rule zero, because it is not really a new rule.

### 1. Standalone pairs

The *pair* of a parent tag and its matching end section tag *as a whole* is standalone if the opening tag clears at the start and the closing tag clears at the end (so that's **BB**, **BE**, **SB** or **SE**). This corresponds to the intuitive notion that such a parent pair should behave like a single standalone partial tag. Another, independent justification of this rule is that the internal delimiters of the pair should not be taken into account because they are in comment space.

Like in standalone tags, the external clearance around a standalone pair should be considered part of the tags rather than part of the literal contents within the surrounding template.

The exact same notion of a standalone pair is also applicable to parameter blocks, because these are sites of template expansion, too. It does not apply to argument blocks, because those are not template expansion sites and also because their outer delimiters are in comment space.

The outermost pair in each of the following examples is standalone:

```
{{<parent}}{{$block}}
{{/block}}{{/parent}}

{{$block}}{{/block}}

    {{$block}}
    {{/block}}
```

### 2. Intrinsic indentation

If the opening tag of an argument block or a standalone parameter block clears at the end (so that's **BB**, **BS**, **BE**, **BN**, **EB**, **ES**, **EE** or **EN** for argument blocks but only **BB** or **BE** for parameter blocks), then the intrinsic indentation of the block is the whitespace at the start of the first line after the opening block tag.

In all remaining cases, the intrinsic indentation of a block is the empty string.

The justification for taking only the opening side of the internal clearance into account is that the absence of start clearance at the closing tag is equivalent to an external partial template that lacks a final line end.

In all examples below, the intrinsic indentation of the block is four spaces:

```
{{$block}}
    content
{{/block}}

{{$block}}
    content{{/block}}

{{<parent}}{{$block}}
    content
{{/block}}{{/parent}}
```

In all examples below, the intrinsic indentation of the block is the empty string:

```
{{$block}}content{{/block}}

    {{$block}}content{{/block}}

    {{$block}}
content
    {{/block}}

    a {{$block}}
    content
    {{/block}}

    {{$block}}
    content
    {{/block}} b

{{<parent}}
    {{$block}}content{{/block}}
{{/parent}}
```

### 3. Deindentation at block definition time

The intrinsic indentation of a block definition is removed from each of its lines before anything else happens with its contents.

The following example block,

```
    {{$block}}
        text
        {{#section}}
            text
        {{/section}}
    {{/block}}
```

is equivalent to an external partial template with the following content.

```
text
{{#section}}
    text
{{/section}}
```

### 4. Indentation at template expansion sites

This rule is a generalization of the indentation rule that already exists for standalone partial tags.

The indentation of a partial tag, parent pair or parameter block pair is determined as follows. If it is a block with intrinsic indentation, then the intrinsic indentation is used. Otherwise, if the tag or pair is standalone, then the indentation of the single tag or the pair's opening tag is used. In all remaining cases, the empty string is used.

Note that the indentation used is always whitespace that is not part of the literal contents of the template.

The indentation thus determined is added to each line of whichever (external) template or block ends up being expanded in the position of the tag or pair, before that template or block is rendered.

In each of the following examples, four spaces of indentation will be added to each line of whichever template or block ends up being expanded on site.

```
    {{>partial}}

    {{<parent}}{{/parent}}

    {{$block}}{{/block}}

{{$block}}
    default content
{{/block}}
```

In each of the following examples, no indentation (the empty string) will be added.

```
{{>partial}}

{{<parent}}{{/parent}}

{{$block}}
{{/block}}
```

## Full circle

Let's revisit some of the examples that appeared earlier in this post, and see how they would render according to the above rules. The playground implements those rules faithfully, so playground savestates are provided.

### The straightforward example

_template.mustache_

```
{{<intermediate}}
{{$greeting}}
high five
{{/greeting}}
{{/intermediate}}
```

_intermediate.mustache_

```
Hi,
    {{<invitation}}{{/invitation}}
```

_invitation.mustache_

```
please give me a:
    {{$greeting}}
    hug
    {{/greeting}}
```

_output_

```
Hi,
    please give me a:
        high five
```

*Try the above example in the [playground](https://jgonggrijp.gitlab.io/wontache/playground.html) by pasting the following code:*

```
{"data":{"text":"null"},"templates":[{"name":"template","text":"{{<intermediate}}\n{{$greeting}}\nhigh five\n{{/greeting}}\n{{/intermediate}}"},{"name":"intermediate","text":"Hi,\n    {{<invitation}}{{/invitation}}"},{"name":"invitation","text":"please give me a:\n    {{$greeting}}\n    hug\n    {{/greeting}}"}]}
```

### The complex indentation example

_template.mustache_

```
{{<intermediate}}
    {{$greeting}}
        high five
    {{/greeting}}
{{/intermediate}}
```

_invitation.mustache_

```
please give me a:
    {{$greeting}}
        hug
    {{/greeting}}
```

_output_

```
Hi,
    please give me a:
            high five
```

*Try the above example in the [playground](https://jgonggrijp.gitlab.io/wontache/playground.html) by pasting the following code:*

```
{"data":{"text":"null"},"templates":[{"name":"template","text":"{{<intermediate}}\n    {{$greeting}}\n        high five\n    {{/greeting}}\n{{/intermediate}}"},{"name":"intermediate","text":"Hi,\n    {{<invitation}}{{/invitation}}"},{"name":"invitation","text":"please give me a:\n    {{$greeting}}\n        hug\n    {{/greeting}}"}]}
```

### The mixed clearance example

_template.mustache_

```
{{<intermediate}}{{$greeting}}
high five
{{/greeting}}{{/intermediate}}
```

_invitation.mustache_

```
please give me a:
    {{$greeting}}hug{{/greeting}}
```

_output_

```
Hi,
    please give me a:
        high five
```

*Try the above example in the [playground](https://jgonggrijp.gitlab.io/wontache/playground.html) by pasting the following code:*

```
{"data":{"text":"null"},"templates":[{"name":"template","text":"{{<intermediate}}{{$greeting}}\nhigh five\n{{/greeting}}{{/intermediate}}"},{"name":"intermediate","text":"Hi,\n    {{<invitation}}{{/invitation}}"},{"name":"invitation","text":"please give me a:\n    {{$greeting}}hug{{/greeting}}"}]}
```

## Discussion

What do you think? Can you think of edge cases where the proposed rules would give undesirable results? Can you think of rules that would give acceptable results in more cases, or simpler rules that would give equally acceptable results?

I hereby offer to write the specs if we reach an agreement. In that case, I will try them with my own implementation before submitting a pull request.

CC to all implementers and contributors who have been recently active around the spec (in random order): @gasche @softmoth @adam-fowler @Danappelxx @pvande @splumhoff @bobthecow @sayrer @spullara

jgonggrijp:
When we pass a block to a parent, we would ideally expect it to "magically" adjust its indentation.

*template.mustache*

```
{{<parent}}
        {{$block}}
        Hello
        {{/block}}
{{/parent}}
```

*parent.mustache*

```
I say,
  {{$block}}
  Hi there!
  {{/block}}
```

*output*

```
I say,
  Hello
```

*Try the above example in the [playground](https://jgonggrijp.gitlab.io/wontache/playground.html) by pasting the following code:*

```json
{"data":{"text":"null"},"templates":[{"name":"template","text":"{{<parent}}\n        {{$block}}\n        Hello\n        {{/block}}\n{{/parent}}"},{"name":"parent","text":"I say,\n  {{$block}}\n  Hi there!\n  {{/block}}"}]}
```

I wrote an extremely long post in #130 on how this might be approached, inviting for a discussion. However, I realized that the post was so long that most people simply might not find the time (or patience or courage) to read it. I hope this PR will be an acceptable alternative to those people, being more concrete and containing substantially less text.

For a detailed rationale behind the specs that I'm proposing here, please refer to #130. The short version is that I applied the following rules. I was able to fully conform my own implementation to these specs. I needed to grow the code size by only seven lines on balance, though there were some subtleties that I'll mention in code comments below.

0. **Standalone tags** (pre-existing rule). Individual block and parent tags should not introduce blank lines in the output when they clear on both delimiters.
1. **Standalone pairs.** The *pair* of a parent tag and its matching end section tag *as a whole* is standalone if the opening tag clears at the start and the closing tag clears at the end. This notion of a standalone pair is also applicable to parameter blocks, because these are sites of template expansion, too. It does not apply to argument blocks, because those are not template expansion sites.
2. **Intrinsic indentation.** If the opening tag of an argument block or a standalone parameter block clears at the end, then the intrinsic indentation of the block is the whitespace at the start of the first line after the opening block tag. In all remaining cases, the intrinsic indentation of a block is the empty string.
3. **Deindentation at block definition time.** The intrinsic indentation of a block definition is removed from each of its lines before anything else happens with its contents.
4. **Indentation at template expansion sites** (generalization of existing standalone partial indentation rule). The indentation of a partial tag, parent pair or parameter block pair is determined as follows. If it is a block with intrinsic indentation, then the intrinsic indentation is used. Otherwise, if the tag or pair is standalone, then the indentation of the single tag or the pair's opening tag is used. In all remaining cases, the empty string is used. The indentation thus determined is added to each line of whichever (external) template or block ends up being expanded in the position of the tag or pair, before that template or block is rendered.

Merging this PR would close #130.

CC to all people I CC'ed in #130: @gasche @softmoth @adam-fowler @Danappelxx @pvande @splumhoff @bobthecow @sayrer @spullara .

### DIFF
--- a/specs/~inheritance.yml
+++ b/specs/~inheritance.yml
@@ -304,3 +304,23 @@ tests:
       Hi,
         one
         two
+
+  - name: Nested block reindentation
+    desc: Nested blocks are reindented relative to the surrounding block
+    data: {}
+    template: |
+      {{<parent}}{{$nested}}
+      three
+      {{/nested}}{{/parent}}
+    partials:
+      parent: |
+        {{<grandparent}}{{$block}}
+          one
+          {{$nested}}
+            two
+          {{/nested}}
+        {{/block}}{{/grandparent}}
+      grandparent: "{{$block}}default{{/block}}"
+    expected: |
+      one
+        three
--- a/specs/~inheritance.yml
+++ b/specs/~inheritance.yml
@@ -235,3 +235,72 @@ tests:
     partials:
       parent: "{{#nested}}{{$block}}You say {{fruit}}.{{/block}}{{/nested}}"
     expected: I say bananas.
+
+  - name: Standalone parent
+    desc: A parent's opening and closing tags need not be on separate lines in order to be standalone
+    data: {}
+    template: |
+      Hi,
+        {{<parent}}{{/parent}}
+    partials:
+      parent: |
+        one
+        two
+    expected: |
+      Hi,
+        one
+        two
+
+  - name: Standalone block
+    desc: A block's opening and closing tags need not be on separate lines in order to be standalone
+    data: {}
+    template: |
+      {{<parent}}{{$block}}
+      one
+      two{{/block}}
+      {{/parent}}
+    partials:
+      parent: |
+        Hi,
+          {{$block}}{{/block}}
+    expected: |
+      Hi,
+        one
+        two
+
+  - name: Block reindentation
+    desc: Block indentation is removed at the site of definition and added at the site of expansion
+    data: {}
+    template: |
+      {{<parent}}{{$block}}
+          one
+          two
+      {{/block}}{{/parent}}
+    partials:
+      parent: |
+        Hi,
+          {{$block}}
+          {{/block}}
+    expected: |
+      Hi,
+        one
+        two
+
+  - name: Intrinsic indentation
+    desc: When the block opening tag is standalone, indentation is determined by default content
+    data: {}
+    template: |
+      {{<parent}}{{$block}}
+      one
+      two
+      {{/block}}{{/parent}}
+    partials:
+      parent: |
+        Hi,
+        {{$block}}
+          default
+        {{/block}}
+    expected: |
+      Hi,
+        one
+        two
--- a/specs/~inheritance.json
+++ b/specs/~inheritance.json
@@ -245,6 +245,62 @@
         "parent": "{{#nested}}{{$block}}You say {{fruit}}.{{/block}}{{/nested}}"
       },
       "expected": "I say bananas."
+    },
+    {
+      "name": "Standalone parent",
+      "desc": "A parent's opening and closing tags need not be on separate lines in order to be standalone",
+      "data": {
+      },
+      "template": "Hi,\n  {{<parent}}{{/parent}}\n",
+      "partials": {
+        "parent": "one\ntwo\n"
+      },
+      "expected": "Hi,\n  one\n  two\n"
+    },
+    {
+      "name": "Standalone block",
+      "desc": "A block's opening and closing tags need not be on separate lines in order to be standalone",
+      "data": {
+      },
+      "template": "{{<parent}}{{$block}}\none\ntwo{{/block}}\n{{/parent}}\n",
+      "partials": {
+        "parent": "Hi,\n  {{$block}}{{/block}}\n"
+      },
+      "expected": "Hi,\n  one\n  two\n"
+    },
+    {
+      "name": "Block reindentation",
+      "desc": "Block indentation is removed at the site of definition and added at the site of expansion",
+      "data": {
+      },
+      "template": "{{<parent}}{{$block}}\n    one\n    two\n{{/block}}{{/parent}}\n",
+      "partials": {
+        "parent": "Hi,\n  {{$block}}\n  {{/block}}\n"
+      },
+      "expected": "Hi,\n  one\n  two\n"
+    },
+    {
+      "name": "Intrinsic indentation",
+      "desc": "When the block opening tag is standalone, indentation is determined by default content",
+      "data": {
+      },
+      "template": "{{<parent}}{{$block}}\none\ntwo\n{{/block}}{{/parent}}\n",
+      "partials": {
+        "parent": "Hi,\n{{$block}}\n  default\n{{/block}}\n"
+      },
+      "expected": "Hi,\n  one\n  two\n"
+    },
+    {
+      "name": "Nested block reindentation",
+      "desc": "Nested blocks are reindented relative to the surrounding block",
+      "data": {
+      },
+      "template": "{{<parent}}{{$nested}}\nthree\n{{/nested}}{{/parent}}\n",
+      "partials": {
+        "parent": "{{<grandparent}}{{$block}}\n  one\n  {{$nested}}\n    two\n  {{/nested}}\n{{/block}}{{/grandparent}}\n",
+        "grandparent": "{{$block}}default{{/block}}"
+      },
+      "expected": "one\n  three\n"
     }
   ]
 }


specs/~inheritance.yml
```diff
@@ -224,3 +224,92 @@ tests:
     partials:
       parent: "{{$foo}}default content{{/foo}}"
     expected: default content
+
+  - name: Standalone parent
+    desc: A parent's opening and closing tags need not be on separate lines in order to be standalone
```
jgonggrijp, line 229: That is, this pair of tags behaves exactly like a single standalone partial tag.

specs/~inheritance.yml
```diff
@@ -224,3 +224,92 @@ tests:
...
+      parent: |
+        one
+        two
+    expected: |
+      Hi,
+        one
+        two
+
+  - name: Standalone block
+    desc: A block's opening and closing tags need not be on separate lines in order to be standalone
```
jgonggrijp, line 244: This is however not exactly the same as a single standalone partial tag, because that would conflict with the existing "Inherit indentation" spec on lines 154-164. "Standalone" here mostly refers to whether the block as a whole will be indented. Contrary to the previous spec, if the end section tag is not standalone by itself, then the trailing newline will **not** be removed from the output if the block ends up being empty.

specs/~inheritance.yml
```diff
@@ -224,3 +224,92 @@ tests:
...
+    data: {}
+    template: |
+      {{<parent}}{{$block}}
+      one
+      two{{/block}}
```
jgonggrijp, line 249: Note this little tweak I had to make here. If I had put the `{{/block}}` at the start of the next line instead, I would have to include double newlines at the end of the expected output, for consistency with the pre-existing "Inherit indentation" spec.

specs/~inheritance.yml
```diff
@@ -224,3 +224,92 @@ tests:
...
+    template: |
+      {{<parent}}{{$block}}
+          one
+          two
+      {{/block}}{{/parent}}
+    partials:
+      parent: |
+        Hi,
+          {{$block}}
+          {{/block}}
```
jgonggrijp, line 272: Here, I'm avoiding the double newline problem by making the block tag and the end section tag standalone by themselves, so their trailing newlines are removed from the output.

spullara:
I am mostly against all this stuff. It requires you to reformat the output which means you have an additional layer of buffering in many cases. For performance reasons I have elided some of the whitespace constraints because of it. Imagine it for recursive includes and the like, as opposed to the simple case you could do at compile time.

gasche:
@spullara My intuition would be that the performance situation is no worse than with the existing rule on partials indentation: you have to track the current "line indentation" and process user-provided input (whenever the current indentation is non-empty) to add this indentation after newlines. This can be done in a zero-copy way by processing the input line-by-line, outputting indentation in between -- or indeed by using a temporary buffer that is output as a whole afterwards.

jgonggrijp:
@spullara I'm not doing any buffering (not even a single layer) and all deindentation happens at compile time. My implementation passes all the specs without doing anything new at rendering time, except that indentation is now not only applied to partials and parents but also to blocks.

I *think* my implementation will also handle recursive includes transparently, though I haven't tested that yet. Could you give an example of a spec that you think would be problematic?

@gasche The partials spec actually states that only the template is indented, not the user input. This is also what I have implemented for parents and blocks. It's not as pretty as indenting the final output, but substantially easier and more efficient to implement.

https://github.com/mustache/spec/blob/bb63070e701e6e5ba9fa0b4adaa259a0ef8115be/specs/partials.yml#L82-L100

jgonggrijp:
Update: in the meanwhile, I open-sourced my implementation, [Wontache](https://gitlab.com/jgonggrijp/wontache/-/blob/integration/mustache/mustache.coffee), so I can show you how I implemented the proposed spec (as well as 100% of the latest release of the actual spec, plus the proposed dynamic partials per #134). Blocks, like partials/parents, are represented as separate templates in Wontache. Compiling and rendering a template happens in four stages. I'm sure many Mustache implementations work like this, but I'll list them just to be clear:

1. Tokenization
2. Parsing
3. Code generation (JavaScript functions with a runtime dependency on support code in Wontache)
4. Template execution (which sometimes involves recompiling a template)

In Wontache, the removal of intrinsic indentation from a block happens during code generation, over here:

https://gitlab.com/jgonggrijp/wontache/-/blob/integration/mustache/mustache.coffee#L403-441

Indentation may be added (back) during template execution, depending on the context in which the block, partial or parent is expanded. In that case, it is recompiled with added indentation. The recompiled template is cached, so this needs to be done only once for each level of indentation encountered. It's a bit crude, but fast. This "magic" happens over here:

https://gitlab.com/jgonggrijp/wontache/-/blob/integration/mustache/mustache.coffee#L224-266

@spullara While it may not be immediately obvious, the above code proves that the output is not reformatted and that no additional buffering is needed. Apart from infrequent template recompilation, it's just string concatenation, immediately followed by output.

Feedback welcome!

agentgt:
Hi folks. I help maintain handlebars.java and somewhat JMustache.


I also have a [Java static mustache implementation](https://github.com/snaphop/static-mustache) in the works which is a fork of another project someone else started (this point will become relevant).  It passes the spec with the exception of #139 . 

I also think the block contents should be eagerly executed: https://github.com/mustache/spec/blob/5d3b58ea35ae309c40d7a8111bfedc4c5bcd43a6/specs/~inheritance.yml#L228-L237

(in the above apples should show and not bananas in my implementation). However I will save that rant for another time.


Let me first say this spec provides invaluable test cases (that I didn't have to write :)) and I am much appreciative of it however:

I have used Mustache and Handlebars in production settings for over 10 years with massive code bases. I have some idea of what works, what people care about, what they want and most importantly what they do not care about.

Whitespace is something people do not care as much about as the spec would lead you to believe especially in the context of generating HTML. 

In fact over and over and over bugs have been filed as folks already have a hard time understanding standalone tags *because the time people care about it they want to manage it explicitly which most templating languages do by default!*. Usually this is when they are generating something other than HTML.

So there are two groups that I see using Mustache:

 * People generating HTML. They don't view source anymore but use their browsers web developers tools. They don't care about white space. Actually if anything these folks care about performance per @spullara point.
 * People generating actual source code. Lots of the Open API tools use Mustache and or Handlebars. These folks would rather standalone and whatever auto whitespace we do did not exist.

Thus given the way the spec is written currently I think block tags should be treated as explicit (that is whitespace between the ($) tags is exactly preserved). It is the easiest addendum to the spec.

The exception of them having standalone semantics *just like the current semantics for sections* I would also be OK with.... but this whole pairing of tags and special semantics for blocks while may only take 7 lines in javascript is absolutely nontrivial for my Java static mustache (particularly because I did not write the tokenizer) and I can't imagine the complexity as well as perf (to @spullara point) when dealing with dynamic lambdas.

Instead of all this continued focus on whitespace:

* I rather the spec have way more test cases (for example there is not a single test case for multiple sibling parents in a single template). 
* I rather lambdas be more powerful (and in someways the could already be with the blocks)
*  What people want is function calls like handlebars helpers. They don't want the incredibly confusing dynamic scoping of inheritance (the parent template can change the context... ask yourself how many programming languages allow this macro like expansion and how rife with bugs they are.... people want lexical scoped functions).

Anyway I was hoping to be 100% compliant with my implementation but I don't feel it is worth the effort for whitespace given my experience and especially so if it keeps getting more complicated.

bobthecow:
@agentgt on my way out the door for a long weekend in the woods, so i haven't spent the time forming an opinion on your issue, but i do know that most implementations that diverge from the spec do it on this issue, so you'd be in good company :)

agentgt:
> However, I can confirm from my own observations that implementers have a hard time wrapping their heads around how to implement it correctly. I suspect that especially incorrect implementations lead to bug reports.

There is not a single Java implementation that implements the whitespace rules correctly (my implementation will be the closest). The whitespace rules are currently more complicated than the whitespace rules for XML normalization. If the developers can't keep the model in their head it is no wonder end users that care about whitespace have similar problems.  

> Could you provide some examples?

https://github.com/samskivert/jmustache/issues/126

I have seen many more usually with code generation. More often than not the request is to remove whitespace handling (ie no standalone) once the user figures out what is going on.

> Me too. Please feel welcome to join the discussion in https://github.com/mustache/spec/issues/135.

I did.

> I don't. Mustache was conceived as a simple, logic-free yet expressive template language. I appreciate that concept and would like to keep it that way, even if that means that I occasionally need to take some extra steps in data preparation on the side of the programming language. From earlier discussions in this issue tracker, I know that I am not the only one with that opinion.

Yet we have lambdas aka functions embedded in JSON, we have dynamic creation of templates produced by lambdas as well as recursive execution of templates as well as dynamic scoping. I wouldn't even be surprised if some implementations are turing complete.

What people are asking for is context object + another parameter -> string. I have provided hosted CMS like systems for some time (e.g. user templates) and it is a lot easier to white list a couple of function calls then prevent all the other possibly behavior you can do with mustache (mustache.java even mentions this).


> Isn't that a conscious design choice in Mustache, though? Simplicity over bells and whistles? If I wanted lots of bells and whistles to fine-tune the whitespace in my templates, **I would probably use ERB or Jinja instead of Mustache**. (That said, you can gain explicit control over whitespace in Mustache if needed, by moving newlines inside tags.)

Because the end user often does not have a choice to pick their templating. Tons and tons of Java libraries have chosen Mustache or Handlebars to do code generation. 

Or in my case I'm coming from knowing 50 or so of offshore not-very-techy developers who use our products handlebars/mustache to generate HTML, CSS, etc. These guys do not care about whitespace but what I did get repeatedly asked is for basic function calls... otherwise they would dip into using Javascript.  That is why we use a locked down Handlebars for a good portion of our services.



> The way I understand the spec (and implemented it), block scope is already lexical. However, you should not think of a block as a closure over the template in which it is defined, but as an independent template with its own root context. Consider this: if a block would close over the template of origin, the destination template would never be able to pass any values into it.

Maybe because it is Friday and I'm exhausted but I'm not sure I follow but I'm talking about parents changing the context.

> The right place for that future rant is in https://github.com/mustache/spec/pull/125 (beginning of discussion) and https://github.com/mustache/spec/pull/129 (remainder). Do keep in mind that the decision was already taken, though, and that multiple people agreed with it.

The parent changing the context scope IMO is dynamic scoping or you could call it just plain macro expansion. I can implement it that way and I probably begrudgingly will but I think it is confusing for most. It is ripe for bugs but I suppose I can make that configurable and ditto for whitespace compliance.


> If you'd appreciate it, I'm willing to take a look at your code and try whether I can make any suggestions that might help you get closer to the spec.

Feel free to take a look: https://github.com/snaphop/static-mustache/blob/snaphop/static-mustache-apt/src/main/java/com/snaphop/staticmustache/apt/TemplateCompiler.java

I will warn you the code is awful..ly complicated. I inherited the project and would have written the tokenizer different. It is a lot more complicated than most Mustache implementations because I have two trees to deal with: the mustache code, and the Java AST (via the annotation processor) as well of course generating Java code which mind you is far more complicated than Javascript because of types. 

However I did put the whitespace logic in as the original author did not have it in so I take fault that it is overly complicated. I was hoping not to read and buffer all tokens till new line but given what apparently I need to do I should have done that all along and is probably why it is complicated.

Oh and maybe I'm just dumb but I still really do not understand why the newline remains at the end on #139 but I am trying as I said I hope to be one of the most compliant and typesafe to boot (well provided you don't just use maps aka hashes).

jgonggrijp:
> [samskivert/jmustache#126](https://github.com/samskivert/jmustache/issues/126)

Could you give another example? In this one, the user expects exactly what the spec prescribes (i.e., the newline should be preserved).

> Yet we have lambdas aka functions embedded in JSON, we have dynamic creation of templates produced by lambdas as well as recursive execution of templates as well as dynamic scoping. I wouldn't even be surprised if some implementations are turing complete.

I meant *syntactically* simple. You are right to point out that there is an explosion of possibility at the semantic level. This is also what I meant by "expressive".

> 
> What people are asking for is context object + another parameter -> string. I have provided hosted CMS like systems for some time (e.g. user templates) and it is a lot easier to white list a couple of function calls then prevent all the other possibly behavior you can do with mustache (mustache.java even mentions this).

I cannot really comment on this part. Maybe Mustache is not the best tool for this particular situation; I'll readily concede that Mustache is not a catch-all solution.

> That is why we use a locked down Handlebars for a good portion of our services.

Why not use Handlebars everywhere?

Side note: Handlebars retains the whitespace rules of Mustache (if you don't use the added delimiter modifiers to adjust how whitespace is handled).

> 
> > The way I understand the spec (and implemented it), block scope is already lexical. However, you should not think of a block as a closure over the template in which it is defined, but as an independent template with its own root context. Consider this: if a block would close over the template of origin, the destination template would never be able to pass any values into it.
> 
> Maybe because it is Friday and I'm exhausted but I'm not sure I follow but I'm talking about parents changing the context.

Yes. I think you need a different mental model. Nothing is "changing" the context. Rather, the parent *template* (i.e., **not** the parent *tag pair*) is *providing* the context to the block *at the site of interpolation* (you can think of "providing context" as passing an argument to a function). Equivalently, you can imagine that the content of the parent template is substituted for the content of the parent tag pair, overriding blocks as provided. To illustrate:

*parent.mustache*

```mustache
{{#people}}
  {{! site of interpolation}}
* {{$greeting}}Hello, {{name}}{{/greeting}}
{{/people}}
```

*template.mustache*

```mustache
{{<parent}}
    {{! site of definition}}
    {{$greeting}}How do you do, {{name}}?{{/greeting}}
{{/parent}}
```

Rendering `template.mustache` above is equivalent to rendering a single template that looks like this:

```mustache
{{#people}}
* How do you do, {{name}}?
{{/people}}
```

Imagine passing the following chunk of data to either template. You'll find that the spec also predicts the same result in both cases. It's just lexical scoping.

```json
{
    "people": [
        {"name": "Fred"},
        {"name": "George"}
    ]
}
```

```
* How do you do, Fred?
* How do you do, George?
```

If I understood your previous comment correctly, you seemed to propose that the `{{name}}` should already be resolved in `template.mustache`, in the current context of the `{{<parent}}{{/parent}}` pair. Inheritance could never work as intended in that case; this example would render as a list with two incomplete greetings, since `name` is not defined in that context.

```
* How do you do, ?
* How do you do, ?
```

> The parent changing the context scope IMO is dynamic scoping or you could call it just plain macro expansion. I can implement it that way and I probably begrudgingly will (...)

Please don't! Such an implementation would be based on a misunderstanding. And yes, that would be confusing and a recipe for bugs as you wrote next. It is better to do something that you thoroughly understand, than to follow some authoritative rule that doesn't make sense to you.

> Feel free to take a look: https://github.com/snaphop/static-mustache/blob/snaphop/static-mustache-apt/src/main/java/com/snaphop/staticmustache/apt/TemplateCompiler.java
> 
> I will warn you the code is awful..ly complicated. I inherited the project and would have written the tokenizer different. It is a lot more complicated than most Mustache implementations because I have two trees to deal with: the mustache code, and the Java AST (via the annotation processor) as well of course generating Java code which mind you is far more complicated than Javascript because of types.

Thanks! I'll study the code, this might take a while. I will not judge you or your code.

> (...) I was hoping not to read and buffer all tokens till new line but given what apparently I need to do I should have done that all along and is probably why it is complicated.

I'm not entirely sure what you mean by "buffering" here, but as I have written before, I don't think it should be necessary.

My implementation keeps hold of the raw template text until tokenization and parsing are completed. Other than the raw text, the token stream and the parse tree, no intermediate strings, copies, lookups or whatever are used anywhere. Both code generation and rendering are just straight-to-output string concatenation.

> 
> Oh and maybe I'm just dumb but I still really do not understand why the newline remains at the end on #139

I'll try whether I can clarify matters for you over there.

> but I am trying as I said I hope to be one of the most compliant and typesafe to boot (well provided you don't just use maps aka hashes).

That's a noble goal.

agentgt:
@jgonggrijp Alright after getting some sleep and playing with Wontache I understand how it is supposed to work. Block tags now are inherently special standalone (e.g. rule 1) in a parent template (ie not surrounded by `{{<`) and parent tags in the child template are special which is exactly how you described and I was just too dense yesterday to comprehend it.

I can implement it but it will require multiple passes (or replaying the tokens) which I was avoiding. 

What I meant before about where I said I was OK with blocks being standalone was for the original standalone behavior (you know where you said I was being contradictive). I know see the confusion.

I'm mostly fine with this "special" standalone behavior but it complicates parsing.

I also don't know what is the correct behavior for embedded block tags (ie block tags within block tags) but that is sort of tangental to the spacing problem.

For example:

```
asdf asdf {{$b1}}{{$b2}}
hello
{{/b2}}{{/b1}}
```

Assuming no replacement would be:

```
asdf asdf hello
```

So b1 and b2 are standalone even though b1 has a tag immediately following it. I am not sure why I find that confusing but I do but if I think about it as a template expansion site it sort of makes more since.


Oh btw I'm a massive hypocrite on the code generation not wanting automatic whitespace handling as I actually used JMustache over mustache.java to generate the spec test code.

jgonggrijp:
> @jgonggrijp Alright after getting some sleep and playing with Wontache I understand how it is supposed to work.

Hooray! Although I'm afraid there is a bug in Wontache which might still have confused you somewhat. Wontache mostly does the right thing, but you cleverly exposed a bug that goes unnoticed by the existing tests in the spec.

By all means, do use Wontache as an example of how the spec might be implemented, but please keep in mind that it might sometimes still be wrong.

> Block tags now are inherently special standalone (e.g. rule 1) in a parent template (ie not surrounded by `{{<`)

Yes, although that *should* not be the case if there is something preceding the opening tag, or following the closing tag.

> I'm mostly fine with this "special" standalone behavior but it complicates parsing.

In the parsing stage, I think it's just a matter of keeping track of clearance before and after the tag. If you store that information in the parsing tree, the clearance of matching tags is easy to retrieve when you need it. To be honest, I think the code generation stage is more complicated (as far as this feature is concerned), especially if you implement reindentation.

> 
> I also don't know what is the correct behavior for embedded block tags (ie block tags within block tags) but that is sort of tangental to the spacing problem.
> 
> For example:
> 
> ```
> asdf asdf {{$b1}}{{$b2}}
> hello
> {{/b2}}{{/b1}}
> ```
> 
> Assuming no replacement would be:
> 
> ```
> asdf asdf hello
> ```


It *should* be

```
asdf asdf
hello
```

but Wontache is doing it wrong. Just like in #139, you found a bug in Wontache.

And Wontache is probably doing it wrong exactly because of the spacing rules, so I guess it's not entirely tangential. Although I still have to look into the exact reason. I will report back on that in #139 as I also announced there.

> 
> So b1 and b2 are standalone even though b1 has a tag immediately following it. I am not sure why I find that confusing

Because you are right and Wontache is wrong! There, you can hang that acknowledgement on your fridge or something. :-)

> Oh btw I'm a massive hypocrite on the code generation not wanting automatic whitespace handling as I actually used JMustache over mustache.java to generate the spec test code.

And I presume JMustache does more of that whitespace management than mustache.java? Let us just enjoy the irony of that.

jgonggrijp:
Like in #130, I have updated my posts to include links to and savestates for my newly developed [Mustache playground](https://jgonggrijp.gitlab.io/wontache/playground.html). I hope this will reduce barriers for readers to experiment with the examples.

The playground is powered by my own implementation, Wontache. Because of that, it fully implements the latest spec as well as the rules proposed here. It aims to be a general, more powerful alternative to the "[demo](https://mustache.github.io/#demo)" on the official Mustache website. Whitespace is handled correctly, lambdas are supported and it is possible to enter multiple named templates alongside each other, so that partial and parent tags can be used effectively. And as demonstrated above, it is possible to share sessions with other people by copy-pasting chunks of JSON code. Please feel welcome to use it.

FYI @agentgt @spullara @bobthecow. The bug that @agentgt discovered (discussed above) has been fixed in the latest release of Wontache and it is not present in the playground.

jgonggrijp:
@gasche Have your thoughts on the matter evolved in the meanwhile? Would you approve?

I decided to draw your attention again, because I got in touch with @spullara and he told me not to take his previous response as a full rejection.

CC @bobthecow: what are your thoughts on this one?

agentgt:
FWIW and in great irony given my previous comments on this thread I'm for these changes!

The whitespace handling of Mustache has become my favorite features such that I have even implemented it in extensions (not in spec) like fragments which happens to be similar to parent/blocks: https://jstach.io/jstachio/#mustache_fragments

(fragments are basically referencing a subset of a template as a partial and in this case we use section-like tag pairs as the subset... its actually a pretty cool feature I wish I had come up with but the HTMX guys did).

jgonggrijp:
At this point, if I'm right, all participants in this pull request are somewhere on the spectrum between neutral and in favor:

- I proposed it.
- @spullara initially stated disagreement, but later wrote in private communication that he thought he should let me merge it.
- @gasche expressed agreement both here and in #130.
- @agentgt finished with a comment in favor.
- @bobthecow attached a ❤️ to the latter comment by @agentgt.

I will now rebase the branch, address merge conflicts and then merge it. I will check for other open PRs that could be merged quickly before tagging a new release.

PR #137, opened 2022-03-22
add test: comment content colliding with variable

### LINKED ISSUE #136: Comments: Behavior on variable name collision
Dear Maintainers,

I've noticed that some implementations treat `{{! comments}}` just as normal variable replacements, just hoping that the user does not define a variable `! comments`.

Example:
Data: `{"! blabla": "never shown"}`
Template: `>{{! blabla}}<`
Rendered A: `>never shown<`
Rendered B: `><`

Handling comments as (hopefully) undefined variables yields Rendered A, but I think the spec suggests Rendered B.

Which is correct?

There is no test case ensuring this behavior, can one get added?
(I can hack together a PR, but I'm not certain that I'll get it to run.)

s9105947:
Currently, an implementation treating comments as undefined variables
successfully passes all tests, because both undefined variables and
comments render to empty string.

This clarifies the behavior: A comment MUST NOT render into anything,
under any circumstances. This includes the case where a variable with
the same name as the comment content is defined.

The test data is designed in a way to trigger any possible name
collision, including whitespaces, a leading exclamation mark (!).

Closes #136

### DIFF
--- a/specs/comments.json
+++ b/specs/comments.json
@@ -89,6 +89,18 @@
       },
       "template": "12345 {{! Comment Block! }} 67890",
       "expected": "12345  67890"
+    },
+    {
+      "name": "Variable Name Collision",
+      "desc": "Comments must never render, even if variable with same name exists.",
+      "data": {
+        "! comment": 1,
+        "! comment ": 2,
+        "!comment": 3,
+        "comment": 4
+      },
+      "template": "comments never show: >{{! comment }}<",
+      "expected": "comments never show: ><"
     }
   ]
 }
--- a/specs/comments.yml
+++ b/specs/comments.yml
@@ -101,3 +101,9 @@ tests:
     data: { }
     template: '12345 {{! Comment Block! }} 67890'
     expected: '12345  67890'
+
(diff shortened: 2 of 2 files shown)


jgonggrijp (approved):
The test looks reasonable and thorough to me. 👍 

My own implementation passes without modification.

```
  official mustache specification
    comments.yml
      ✔ ...
      ✔ Variable Name Collision
```

gasche:
This also looks like a good idea to me. Thanks!

jgonggrijp:
@spullara or @Danappelxx I think this one is ready to merge.

jgonggrijp:
Also, I hereby offer to join you as a maintainer, to help process pull requests. No offense if you're not into it (yet).

Danappelxx:
I've actually been trying to find your email to offer you the permissions 😁 my new job doesn't permit me to maintain open source projects so I'd love to hand it off!

jgonggrijp:
That's a happy coincidence! If you managed to find it, please send me an email to tell me how you did it. :-)

Thanks for having me on board.

PR #145, opened 2023-06-02
Added gitattributes

Krinkle:
Upstreaming a trivial change by @jbboehr from https://github.com/jbboehr/mustache-spec/commits/b96be9fd4c6d6984828d93169fe7e86d8a8aec2f, as used at https://github.com/jbboehr/libmustache.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf


jgonggrijp (commented):
Not very familiar with gitattributes, but this looks good at first sight, based on the documentation.

Is there any off chance that this could break anything for anyone?

jbboehr:
Seems like eol=lf might not a good idea but considering I made that commit
almost exactly ten years ago, I have no idea what it was for.

jgonggrijp:
@jbboehr what makes it seem (to you) that eol=lf might not be a good idea?

PR #149, opened 2023-10-31
Fix typo in Go code

jabley:
Also regenerate the JSON from the changed YAML.

Minor correction to #85.


jgonggrijp (approved):
Oops! I should have done that JSON conversion before merging #85, but now I don't need to anymore.

Good thing I didn't issue a new release yet.

PR #195, opened 2025-01-16
Replace an NBSP (ASCII 160) with a space (ASCII 32) in sections.yml

alterationx10:
I ran into an issue parsing the json version of this file on a linux/amd64 test runner (but not my local arm64 Mac). Took me a while to find it, but I believe it's because of this  NBSP (ASCII 160) character.

Most parsers would probably handle this better than the one I made 😁 , but I also thought there wouldn't be any harm in swapping it out for an SPACE (ASCII 32)


jgonggrijp (approved):
Thank you for bringing this to our attention and for contributing a fix.

I reviewed the change with two other programs to triple-check that this is only replacing one NBSP with a regular space. I see no reason why that particular space would need to be a NBSP, so I agree this change is for the better.

If you don't mind, please update the JSON as well. Otherwise, just let me know and I will do it.

alterationx10:
I ran `rake build`, but on my machine it updates the NBSP, but it also modifies every json file (by re-formatting empty objects/arrays to be on one line stead of two), so I opted to remove the NBSP in the json manually, if that's ok.

jjramon1699-ai:
<script src="https://unpkg.com/adaptivecards-designer/dist/adaptivecards-designer.js"></script>

PR #198, opened 2025-04-24
add lambdas for erlang

Eptwalabha:
Adding  the Erlang's version of the lambdas specs. All test cases are covered.

### DIFF
--- a/specs/~lambdas.json
+++ b/specs/~lambdas.json
@@ -17,7 +17,8 @@
           "clojure": "(fn [] \"world\")",
           "lisp": "(lambda () \"world\")",
           "pwsh": "\"world\"",
-          "go": "func() string { return \"world\" }"
+          "go": "func() string { return \"world\" }",
+          "erlang": "fun() -> \"world\" end."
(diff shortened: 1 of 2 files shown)


jgonggrijp (approved):
I don't speak Erlang (yet), but this looks good to me. Thanks for contributing!

I will wait one week for @spullara or @bobthecow to comment. Unless either of them says otherwise, I will merge after that week.

Are you working on an Erlang implementation? In that case, please feel welcome to do a [show and tell](https://github.com/mustache/spec/discussions/categories/show-and-tell).

Eptwalabha:
> I don't speak Erlang (yet), but this looks good to me. Thanks for contributing!  

No problem, I'm the one who should thank you for these specs - they helped me a lot with use cases I hadn't even thought of…

> […] please feel welcome to do a [show and tell](https://github.com/mustache/spec/discussions/categories/show-and-tell).

Sure, I'll do it once the lib reaches v1.0.0 (I still have to implement inheritance)

PR #204, opened 2026-04-25
Implement lambdas for MATLAB

dbuzinski:
Add lambdas for MATLAB to ~lambdas.json and ~lambdas.yml.

The implementations are mostly straightforward, but I had to get a little creative for the "Section" and "Interpolation - Multiple Calls" tests because:
1. [Anonymous functions](https://www.mathworks.com/help/matlab/matlab_prog/anonymous-functions.html) that get evaluated and assigned to a variable cannot contain key words (`if`, `global`, etc.)
2. Anonymous functions must have static workspaces.

Users of a Mustache implementation written in MATLAB won't have to deal with either of these problems because they can write their lambda definitions in an .m file, but I wanted to keep the spec tests as standalone strings.

### DIFF
--- a/specs/~lambdas.json
+++ b/specs/~lambdas.json
@@ -90,7 +90,7 @@
           "pwsh": "if (($null -eq $script:calls) -or ($script:calls -ge 3)){$script:calls=0}; ++$script:calls; $script:calls",
           "go": "func() func() int { g := 0; return func() int { g++; return g } }()",
           "erlang": "fun() -> G=case get(g) of undefined -> 1; _G -> _G end, put(g, G+1), G end.",
-          "matlab": "@()feval(@(x,c)c, evalc('setenv(''COUNT'', string(lookup(dictionary(NaN,1,3,1), str2double(getenv(''COUNT'')), FallbackValue=str2double(getenv(''COUNT''))+1)))'), str2double(getenv('COUNT')))"
+          "matlab": "@() feval(@(~) getappdata(0, 'calls'), evalc('setappdata(0, ''calls'', mod(sum(getappdata(0, ''calls'')), 3) + 1)'))"
         }
       },
       "template": "{{lambda}} == {{{lambda}}} == {{lambda}}",
@@ -137,7 +137,7 @@
           "pwsh": "if ($args[0] -eq \"{{x}}\") {\"yes\"} else {\"no\"}",
           "go": "func(text string) string { if text == \"{{x}}\" { return \"yes\" } else { return \"no\" } }",
           "erlang": "fun(\"{{x}}\") -> \"yes\"; (_) -> \"no\" end.",
-          "matlab": "@(text)lookup(dictionary(true, \"yes\"), text==\"{{x}}\", FallbackValue=\"no\")"
+          "matlab": "@(text) replace(string(text == \"{{x}}\"), [\"false\", \"true\"], [\"no\", \"yes\"])"
         }
       },
       "template": "<{{#lambda}}{{x}}{{/lambda}}>",
@@ -208,7 +208,7 @@
           "pwsh": "\"__$($args[0])__\"",
           "go": "func(text string) string { return \"__\" + text + \"__\" }",
           "erlang": "fun(Text) -> \"__\" ++ Text ++ \"__\" end.",
-          "matlab": "@(text)\"__\" + text + \"__\""
+          "matlab": "@(text) \"__\" + text + \"__\""
         }
       },
       "template": "{{#lambda}}FILE{{/lambda}} != {{#lambda}}LINE{{/lambda}}",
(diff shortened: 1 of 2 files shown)


specs/~lambdas.json
```diff
@@ -86,7 +89,8 @@
           "lisp": "(let ((g 0)) (lambda () (incf g)))",
           "pwsh": "if (($null -eq $script:calls) -or ($script:calls -ge 3)){$script:calls=0}; ++$script:calls; $script:calls",
           "go": "func() func() int { g := 0; return func() int { g++; return g } }()",
-          "erlang": "fun() -> G=case get(g) of undefined -> 1; _G -> _G end, put(g, G+1), G end."
+          "erlang": "fun() -> G=case get(g) of undefined -> 1; _G -> _G end, put(g, G+1), G end.",
+          "matlab": "@() feval(@(~) getappdata(0, \"calls\"), evalc(\"setappdata(0, \"\"calls\"\", mod(sum(getappdata(0, \"\"calls\"\")), 3) + 1)\"))"
```
jgonggrijp, line 93: Please explain how this works.

jgonggrijp (commented):
Thanks for contributing. Please feel welcome to introduce your implementation in a [show and tell](https://github.com/mustache/spec/discussions/categories/show-and-tell).

The lambdas look mostly sound to me, but I need to understand how your most creative solution works before I can approve this. See comment below.

specs/~lambdas.json
dbuzinski: App data is a global state store that is usually intended to share data between callbacks or separate UI figures (see [getappdata](https://www.mathworks.com/help/matlab/ref/getappdata.html) and [setappdata](https://www.mathworks.com/help/matlab/ref/setappdata.html)). I am using it because I don't have access to the keywords `global` or `persistent` and can use only function calls.

My first function wraps a second function and returns its output. I do this so the first function has no inputs. The second function always returns the value stored in app data as "calls".

The second function also takes an unused argument. The unused argument's purpose is to run an `eval` call that updates app data (increments it by 1). I need to use `evalc`, which captures outputs (and returns captured output), rather than `eval` to ensure this produces some output to feval or I get an error.

Looking inside the evalc call, `sum(getappdata(0, "calls"))` this just retrieves the value `calls` from app data. If `calls` hasn't been defined yet, `getappdata` returns an empty array. I wrap it with `sum` to handle the case where `calls` isn't defined so it resolves to `0` (`sum([])` equals `0`). If it's already defined, it resolves to the value itself (`sum(n)` equals `n`). The `mod` is so if I run the tests multiple times, it still produces 1, 2, 3 (I noticed powershell implementation does this too). Then I add 1 and store it back to app data.

A more usual way for an end-user to do this would be to make a file called `count.m` with the following content:
```matlab
function out = count
persistent c;
if isempty(c)
    c = 0;
end
c = c + 1;
out = c;
end
```
Then the user can just reference the lambda as `@count`. Unfortunately, this can't be done if the function is inlined.

Feel free to ask more questions if you have them 😄 It's a workaround because anonymous functions aren't intended to be mutable in this way in MATLAB, but I still wanted to test the behavior.

jgonggrijp: Thanks for the detailed explanation. I think I fully understand what's happening inside the call to `evalc`, but the comma just before the `evalc` is confusing me. Is this a comma expression inside the `@(~)` function, or does that comma separate two arguments to the `feval` call (of which the `@(~)` function would be the first)?

Regardless of the answer to the previous question, why is it the expression *before* the comma that is returned by the outer function, rather than the expression *after* the comma? I think you meant to explain this with the bit about the unused argument, but I cannot wrap my head around how that works yet.

dbuzinski: Good question. What comes before the comma is the definition of the function. What comes after the comma is the argument to the function, which is evaluated before applying.

I am using [`feval`](https://www.mathworks.com/help/matlab/ref/feval.html?s_tid=srchtitle_support_results_1_feval), which takes a function as its first argument and any number of arguments to apply to the function as its remaining arguments. Also `~` is a convention for an unused variable in MATLAB.

I think it's easier to see if I take a few lines to build up the expression. I'll also omit the content from the `evalc` call so it's easier to read.
```matlab
% Inner function.
% It takes one (unused) argument, ~, and returns "calls" from app data.
g = @(~) getappdata(0,"calls");

% To do this in one line, I must use feval.
% The evalc expression gets evaluated first (which updates app data),
% and the output of evalc (some captured command window output string)
% is applied to g as an argument. Current value of "calls" is returned.
feval(g, evalc(expression));

% Outer function.
% It takes no arguments and returns the output of the feval expression.
f = @() feval(g, evalc(expression));
```

In addition to not having access to keywords for inline functions, I can't add a 2nd equals sign, so I can't break up the expression as logically in the string for the spec.

jgonggrijp: OK, got it! Good exlanation. :+1:

